PR #1, opened 2021-03-23
Add signature

ivanka2012:
Normally, I don't support Richard M. Stallman because of his views, but this one's an exception. Social media and cancel culture is a cancer that needs to be eradicated.


nukeop:
Can you link your github profile like I did? Just for reference so it's not just a list of anonymous nicknames

ivanka2012:
You should put this up somewhere in the thread or here

nukeop:
I added it to the readme.

PR #58, opened 2021-03-24
Lynch mobs aren't cool

the-rocinante:
Our society needs strong support for fundamental human rights like freedom of speech and jurisprudence, not mob justice on the basis of smear campaigns. Having reviewed the evidence I feel in good conscience supporting someone (however abrasive of character) who has consistently and doggedly stood for our basic rights and freedom from the very beginning.

https://www.wetheweb.org/post/cancel-we-the-web


nukeop:
Thank you for the support!

PR #162, opened 2021-03-24
Signed

Anti-pin:


README.md
```diff
@@ -4,5 +4,6 @@ To sign, open a pull request. Link your profile or email in the signature. Pleas
 
 Let's keep the tone firm, but professional.
 
+PC went far too far
```
nekonee, line 7: Please add your signature or, if you don't want to support this letter, refrain from changing README.md.

nekonee (changes requested):

README.md
Anti-pin: Fixed

PR #339, opened 2021-03-24
Add myself to the list.

tmtvl:
I will acknowledge the irony of doing this on GH, which RMS wouldn't venture near due to the proprietary JavaScript; but I still hope we can make a difference.


nekonee:
It's the most accessible method.

podvornyakva:
Please. add a file instead of edit index.md

Follow README.md instructions.

Thank you.

PR #459, opened 2021-03-24
Improved signing procedure

cmd410:
This new procedure takes advantage of Jekyll's data files.

New users sign the letter by creating a yaml file in `_data/signed/` with their username and link as follows:

```yaml
name: example
link: https://example.com/
```

This solves the problem of merge conflicts and users accidentally removing other links, as every user will have own data file.

This PR already migrated users from commit f291f94. There is also `migrate.py` script to automatically parse `index.md` file and add users into `_data/signed/`


nukeop:
Thanks, this solves a HUGE pain for us. We're still looking for a good bulk merging solution.

nekonee:
This is great!

PR #608, opened 2021-03-24
Create KOLANICH.yaml

KOLANICH:
Though I dislike and disaggree to some part of the letter (the one saying `You have no part in choosing the leadership of any communities.`, which is absolutely wrong), I'd still sign it.


podvornyakva:
Yep. Human rights is for all. Not just for tall and beautiful.

PR #938, opened 2021-03-24
Signed

ztx-lyghters:


podvornyakva:
Add .yaml extension to file.

ztx-lyghters:
fixed

ztx-lyghters:
@nukeop my signature isn't visible anymore, though .yaml file does exist in __data/singed/_. What happened?

podvornyakva:
Merged to list, probably.

ztx-lyghters:
@podvornyakva I have deleted it after I noticed this problem. There was a message that it can be safely deleted, since it was merged.

> UP: visible again...
Huh? Where exactly?

podvornyakva:
> Huh? Where exactly?

Ahh... I understand what you mean.
Read note:  Sorry, we had to truncate this directory to 1,000 files. 453 entries were omitted from the list. 
Cause your file is at the end of the list, you cant see it. But it viewable in commit.

ztx-lyghters:
> > Huh? Where exactly?
> 
> Ahh... I understand what you mean.
> Read note: Sorry, we had to truncate this directory to 1,000 files. 453 entries were omitted from the list.
> It is a github incapability to show such long list. Cause your file is at the end of the list, you can't see it. But it viewable in commit.
> Instead you can use git clone util and asure that the file is there.

No, this isn't what I meant. My signature isn't showing on THIS page - https://rms-support-letter.github.io
I saw the "trancate" message and I did cloned this repo to check if the .yaml file is still in there, before posting anything. Now read my initial comment about this problem again. I state that .yaml file, in fact, does exist, but my signature is missing (from the site)

ztx-lyghters:
Well, now I'm actually visible. Not sure why, but problem solved, I guess.

PR #999, opened 2021-03-24
Fix yamls

cmd410:
- moved signature yamls from root to `_data/signed`
- removed `<>` in some entries as it is considered a html tag by jekyll
- renamed file that contained non-ascii letter in its name, as it caused errors
- added more clear example


nukeop:
Thanks, I proofread every file but missed some obvious stuff

PR #1059, opened 2021-03-24
Signed

edward-riley:


trufanov-nok:
Hi, plz сheck out the README.md, it seems they want you to create a file with .yaml extension.

PR #1211, opened 2021-03-24
Create AlexanderBasov.yaml

Allexin:


nukeop:
Please only add one link

Allexin:
why?

DOORAK:
> 
> 
> why?

Because link used on main letter page to make signer name clickable, I suppose.

PR #1368, opened 2021-03-24
ivvil

ivvil:


imachug:
The link gives me 'Unable to connect' over HTTPS, is that expected?

cmd410:
Name field should be all lowercase.
Do:
```yaml
name: ...
```
Don't:
```yaml
Name: ...
```
Else it won't be recognized.

PR #1369, opened 2021-03-24
Best wishes to Richard

6r1d:
Thank you for making this support letter. We all know about "Embrace, extend, and extinguish". Richard speaks about freedom of information and often thinks on behalf of the users, which big, faceless corporations don't really like. And we need people like him. The sad thing is to see someone from EFF on the other letter...


imachug:
Thanks for your support!

PR #1380, opened 2021-03-24
[Proof of censorship] rms-open-letter.github.io

HacKurx:
Can an admin of github.com explain me why the bug has been removed? ...
https://github.com/rms-open-letter/rms-open-letter.github.io/issues/1483


nukeop:
Signatures only please

PR #1411, opened 2021-03-24
Signature

mixedCase:
I don't agree on many, many issues with Stallman, but this attack on his persona starts off mischaracterizing his words with malicious intent and therefore I repudiate it.


nukeop:
Please see readme for the format your signature needs to follow. Once you fix the problem, please reopen the PR.

PR #1444, opened 2021-03-24
sign

Kezii:


cmd410:
@nukeop is it ok that the linked organization is literally empty here?

Kezii:
yes it is, since
1) it exists since 2017
2) it's a public name that exists on some platforms and it's owned by me
3) I have not signed with my name, only the org

cmd410:
@Kezii  Oh, okay then, but it kinda looks sus, hope no one creates dummy orgs just to put them on the list.

Kezii:
There you go
immagine

PR #1492, opened 2021-03-25
Create taraak.yaml

Taraake:
Stallman deserves justice and freedom.


Taraake:
Its everything ok with my signature?, I only used git on a small server for my personal writings before and I'm a bit lost.

PR #1515, opened 2021-03-25
Adding my signature and standing up for the principle of Freedom of Speech

robd003:


ChuckSSneed:
Adding my signature. Richard deserves and needs our help.

PR #1531, opened 2021-03-25
Remove my name.

coolshaurya:
I decided to change my opinion about signing after reading this.


imachug:
We're here for truth, not for a big number, so okay, merging

PR #1536, opened 2021-03-25
Added my name and translated index.md to PT-BR, because the open letter had it translated too

MrExcaliburBr:


nukeop:
Thanks, translations help us broaden the reach.

PR #1538, opened 2021-03-25
Rename Signed! to Mapper720.yaml

Mapper720:


imachug:
Done. Thanks for the notice!

PR #1554, opened 2021-03-25
Translated index.md to Spanish

cidoku:
The other letter had a Spanish translation, so I figured we needed one too.


nukeop:
Thanks, this is very valuable.

PR #1557, opened 2021-03-25
A heartfelt thanks to Richard Matthew Stallman

vinniec:
I have always appreciated stallman's attitude of speaking his mind even if it might be uncomfortable or not acceptable to someone.
It was precisely this attitude that made him fight for free software even at the cost of being politically incorrect, in defense of an ideal.

When a person behaves in this way, in an assertive way, he exposes himself to the contrast of people who have different opinions, and this is normal because it stimulates confrontation.
On the other hand, opportunistic people always act in an attempt to stay in the good graces of public opinion, they are ready to say whatever is deemed convenient to improve their position, to show themselves more deserving, even at the cost of deceiving others.

The first type of people expose themselves and run the risk of being wrong, in a healthy environment we should not be afraid of being wrong since we live in a perpetual state of doubt that we try to remedy with trial and error attempts.
The second type of person does not expose themselves, when they make mistakes that harm others they hide behind unassailable dialectical shields, they will deny until death to the point of creating existential problems for those who believe in the good faith of others and question themselves without realizing that the other side does not do the same.

This process is making unhealthy the sociable relationship between people, because being dishonest generates an advantage that the honest ones don't have, and as for the evolutionary law the species with more advantages will be the ones to survive.
And one day we will end up with one of these snakes with forked tongues in charge of ideological associations like the FSF.

Now we're seeing an attack on an honest man by people who don't recognize his work, but instead use some of his controversial opinions to justify the fact that he is not adequate to carry on the movement he founded himself!
Read their opinions and tell me if even one of them has any relation to the right to share knowledge that Stallman has always defended, the size of their hypocrisy is so big that become invisible because if you try to look at it, it comes out of the frame and you can't get a full view of it with all its contours.

### DIFF
--- a/_data/signed/vinniec.yaml
+++ b/_data/signed/vinniec.yaml
@@ -1,0 +1,2 @@
+name: vinniec
+link: https://github.com/vinniec


Bagelboi:
Respect my nigga.

PR #1602, opened 2021-03-25
Create Mostafa Ahangarha

ahangarha:
I wished to make PR in a FLOSSy way. But because of your choices, I got forced to use shithub. Anyway, wish bests for Free Software.


imachug:
We're aware that GitHub is a problem for some people, hopefully we'll be able to set up a mirror soon. Thanks for signing!

ahangarha:
As soon as you make the other repo (which I suggest to be the main one and github become the mirror) I will advertise this petition.

PR #1611, opened 2021-03-25
Rename motor1n to motor1n.yaml

motor1n:


imachug:
I've bulk renamed all invalid files. Thanks for raising the issue!

PR #1628, opened 2021-03-25
Signed template sorts signatures

### LINKED ISSUE #1578: Sort the list?
The list of names is unsorted, so it makes it look a little bit unprofessional. I like sorting it by names as well so that we make sure there are no duplicates.

Can someone take the effort to write a script to sort the list automatically alphabetically?

cmd410:
closes #1578

### DIFF
--- a/_layouts/signed.html
+++ b/_layouts/signed.html
@@ -11,11 +11,19 @@ layout: home
 </p>
 
 <ol>
-    {% assign folder = site.data.signed | sort %}
-    {% for sign_hash in folder %}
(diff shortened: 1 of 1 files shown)


imachug:
Thanks! hopefully it won't break in production :smiley:

imachug:
There are some problems with the generated content: there are spaces in URLs and the first entry is empty. Do you think you can fix that?

image

imachug:
I reverted the merge because of the bugs, I'll merge it back when you or I figure out how to fix them

PR #1633, opened 2021-03-25
Signed by Sergey Kudryashov

SergeyKud:


imachug:
Please add the file to _data/signed directory instead of root

PR #1665, opened 2021-03-25
Create tti-rus.yaml

baterflyrity:
Signed with my russian colleagues from TTI.


imachug:
What link should I set for your name? https://www.tti.com/ ?

baterflyrity:
We dont have separate link so just leave as is.

PR #1705, opened 2021-03-25
Signed the letter and translated it to Turkish

berkan-sahin:
I think having a Turkish version of this letter can help the message reach a wider audience. If Turkish journals pick up on this controversy, most would paint RMS in a bad light, since that's the more popular opinion. Therefore I think it is important to have a Turkish copy of this letter.

### DIFF
--- a/_data/signed/berkan-sahin.yaml
+++ b/_data/signed/berkan-sahin.yaml
@@ -1,0 +1,2 @@
+name: Berkan Sahin
+link: https://bsahin.xyz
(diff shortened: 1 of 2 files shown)


nukeop:
I will merge it, but I don't get your point, do you want to paint RMS in a bad light?

nukeop:
We will also need a link to the Turkish version somewhere on the page

berkan-sahin:
No, of course not. Sorry for the misunderstanding. I meant that if we have a *supportive* letter in Turkish, the press would be more inclined to cover this viewpoint of the controversy rather than the more popular anti-RMS one.

PR #1822, opened 2021-03-25
fix link url

moeryomenko:
add pre-commit hook for validate links.

the previous [pull request](https://github.com/rms-support-letter/rms-support-letter.github.io/pull/1771) made a mistake in the link.

Signed-off-by: Maxim Eryomenko <moeryomenko@gmail.com>

### DIFF
--- a/pre-commit
+++ b/pre-commit
@@ -1,0 +1,38 @@
+#!/bin/bash
+
+function is_valid_url {
+	regex='https?:\/\/(www\.)?[-a-zA-Z0-9@:%._\+~#=]{1,256}\.[a-zA-Z0-9()]{1,6}\b([-a-zA-Z0-9()@:%_\+.~#?&//=]*)'
+	if [[ $1 =~ $regex ]]
+	then
+		exit 0
+	else
+		echo ${BASH_REMATCH[@]}
+		exit 1
+	fi
+}
+
+
+function parse_yaml {
+   local prefix=$2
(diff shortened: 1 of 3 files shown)


pre-commit
```diff
@@ -0,0 +1,33 @@
+#!/bin/sh
```
gorshunovr, line 1: 1. Not everyone works on *nix system. Many users are on Windows or use GitHub UI.

imachug:
I think it'd be easier for users if we ran the check in a GitHub Action. This would help many people who commit via Github UI.

moeryomenko:
> I think it'd be easier for users if we ran the check in a GitHub Action. This would help many people who commit via Github UI.

Maybe in next pull request

pre-commit
moeryomenko: Change README.md, mark this step for *nix users.

imachug:
I'm just afraid that no one will bother to sign the letter if you require them to copy some potentially dangerous executables

moeryomenko:
> if you require them to copy some potentially dangerous executables

Make optional

> some potentially dangerous executables

I disagree, the hook code is visible

imachug:
Well, you know. If people don't care enough to double-check that the file format is valid, they surely won't bother to copy and check some scripts.

moeryomenko:
> Well, you know. If people don't care enough to double-check that the file format is valid, they surely won't bother to copy and check some scripts.

Would agree, it's good point add validation

imachug:
We now have a GitHub Action so closing.

PR #1891, opened 2021-03-25
Add chinese translation.

shenlebantongying:
The Chinese translation is good enough. I am a bilingual and support freedom.


nukeop:
Awesome, thank you!

PR #1912, opened 2021-03-25
Unify translation layouts

cmd410:
Now all translations use the same layout.
Also added link to original English version in translations.


cmd410:
Okay, seems like someone already fixed layouts)

PR #2071, opened 2021-03-25
Italian localization

dag7dev:
I would suggest to squash and merge.


imachug:
I did not squash because I want to keep the entire history and not loose the proof that you committed the signature yourself.

PR #2129, opened 2021-03-25
Added my name

chameleon-lizard:
RMS may be sloppy in his choice of words and I do not endorse his views. However, I think that this outstanding example of frontier justice is not doing opensource any good.


pacogalvan:
Estoy de acuerdo

PR #2224, opened 2021-03-25
Update index_de.md

ggiedke:
various improvement in language:
less nouns, more verbs
propose different wording (verwerflich->bösartig; Angriffe bewerkstelligt -> Angriffe gab, Gremium->Organisation, Purismus->Reinheit; Leiterschaft->Leitung) if more common ones exist
in two places the German text was not the correct translation of the English original!
the original says nothing about whether RMS' undiplomatic way of putting things is useful or not for the free software cause, but only states that he often does not put things diplomatically. Therefore I propose to change "dass seine undiplomatische Ausdrucksweise keine Voraussetzung ist für seinen aktivistischen Erfolg" by "er meist nicht versucht, die Dinge diplomatisch zu formulieren". (I also disagree with the content of the German text; I think RMS' bluntness and clarity *is* partially a prerequisite for his success.) 
And the mob has not attacked him "wider angemessene Argumente" but "aufgrund angemessener Argumente" ...

PS: apologies if it is bad style to make many changes at once; this is the first time I'm proposing changes on git.

### DIFF
--- a/index_de.md
+++ b/index_de.md
@@ -14,17 +14,17 @@ Richard M. Stallman, häufig bekannt unter dem Kürzel RMS, war seit
 Jahrzehnten eine treibende Kraft in der Freien Software-Bewegung,
 mit Beiträgen wie dem GNU Betriebssystem und Emacs.
 
-In jüngster Zeit wurden eine reihe verwerflicher Online-Angriffe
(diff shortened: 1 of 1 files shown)


imachug:
It's fine to propose many changes if they are related. Thanks for the PR!

nukeop:
This is great, thank you! My qualifications to judge quality are limited, but as far as I can tell, it's an improvement.

PR #2268, opened 2021-03-25
Sign

ccorn:
In particular, I oppose the idea of the RMS opponents that *there must be radical change within the community*.


ccorn:
Wrong path, must be in `_data/signed`.

PR #2320, opened 2021-03-25
Adding my signature

1t3x:
long live foss


parnote:
RMS is a champion of FOSS, and should be treated as the treasure he is to the FOSS community, without regard for his own PERSONAL opinions (which we all have). Without RMS, the FOSS community wouldn't be anywhere near as robust as it is. The FOSS community owes an unpayable debt to RMS.

PR #2414, opened 2021-03-26
fixing RTL (hopefully)

danialbehzadi:


imachug:
Thanks, let's try that

PR #2461, opened 2021-03-26
Create elemhunter.yaml

Tenome:
Reinstate Richard M. Stallman.


tdemin:
@Tenome the layout template doesn't check links, and will try to insert "N/A" as a clickable link. Specify https://github.com/Tenome if you don't have a website.

Issue #837, opened 2021-03-24
Make a comment here if you want to be added without making a pull request


nukeop:
Include your name (preferrably real name), and a website or email.
Also acceptable are entries from organizations if you represent them.

appetrosyan:
Good idea!

pizdjuk:
It needs to be referred in the letter

powerelastic:
Add me

pzizzari:
Add me

appetrosyan:
Could you please be more specific: follow the format: 
``` 
name: your real name
url: <your-github-repo>
```

AlexandrSemenuk:
Add me!

appetrosyan:
@AlexandrSemenuk в Интересах удобства, не могли бы вы оставить запрос в формате?

unspecd:
name: Evgeny Ermakov
link: https://github.com/unspecd

i-just-lizard:
```
name: Nikita Manzin
link: https://github.com/i-just-lizard
```

Michaoil:
Add me, thx mate

IIIT0II0P:
name: Gennadiy Toporkov
link: https://github.com/IIIT0II0P

Lortegap:
name: Luis Ortega
link https://github.com/lortegap

appetrosyan:
```
name: Adrien Bourmault
link: neox@os-k.eu
```

LLIAJIYH:
`name: Tagiev Bayram
link: https://github.com/LLIAJIYH
`

pastk:
name: Konstantin Pastbin
link: https://github.com/pastk

ojoeth:
name: Joe Hill
link: joe@joeth.uk

VitalyBykhalov:
I express my support to Stallman.
It's a shame that the community is going crazy because of gender hysteria.

mevsme:
name: Ollie
link: github.com/mevsme

birdboy2000:
name: birdboy2000
link: b1rdboy2000@msn.com

Not a programmer, my support is 100% based on my anti-capitalist politics and this has all the markings of a corporate smear campaign.

davisi3:
name: Davis

unbearableDarkness:
name: Carlo
link: https://github.com/unbearableDarkness

epistemex:
```
name: Ken Fyrstenberg
link: https://github.com/epistemex
```

[...]

PR #2500, opened 2021-03-26
Add signature

kapilavashtu:


6r1d:
Please check the signature format we are using in a readme

imachug:
Hi, please create a file in directory _data/signed or add a comment here: #837.

PR #2519, opened 2021-03-26
Defend RMS

TheGreatGman:


trufanov-nok:
I guess the extension should be yaml, not yml

TheGreatGman:
> I guess the extension should be yaml, not yml

Thanks

PR #2538, opened 2021-03-26
Cancel the Cancel Culture

stanier:
I was on the edge about signing simply because there is not much obvious evidence for or against him at this time.  The RMS ~~Open~~ Cancel Letter is intentionally vague about the details and would much rather call him all the trigger words they know to paint someone as the boogie-man.

Let's not let them cancel.

Also the cancel letter is on github.com/trending, but the support letter isn't despite similar activity...

Also, who *actually* thinks they can somehow get the entire board of an organization to resign over harmless debate ramblings of one individual.  That is the stupidest thing I have ever heard of.

### DIFF
--- a/_data/signed/stanier.yaml
+++ b/_data/signed/stanier.yaml
@@ -1,0 +1,2 @@
+name: Keyton Stanier
+link: https://github.com/stanier


rkfg:
If you choose "this week" or "this month" on the trending page, you wouldn't even see that repo there. It was _clearly_ pinned manually, the number of stars is even lower than here as well as the number of contributions. As usual, _they_ can't win a fair game so they'll always cheat.

rkfg:
Update: this repo is now at the top in trending! Must be a coincidence of course...

PR #2598, opened 2021-03-26
Add the option for mailing in support

Tyil:
I've created a mailing list on [Sourcehut](sr.ht), to allow people to mail in their signatures. I'm not great at every language, so I can't update all the different language posts.

### DIFF
--- a/index.md
+++ b/index.md
@@ -67,4 +67,6 @@ Especially not via another mob attack which does not remotely
 resemble a fairly conducted debate as exemplified 
 by better people such as Richard Stallman.
 
-To sign, please [submit a pull request](https://github.com/rms-support-letter/rms-support-letter.github.io/pulls).
+To sign, please [submit a pull
+request](https://github.com/rms-support-letter/rms-support-letter.github.io/pulls),
+or [send a signed patch](mailto:~tyil/rms-support@lists.sr.ht).


nukeop:
How can we access those emails?

Tyil:
It's a public mailing list, you can subscribe and receive them. If nobody else wants to be subscribed, I can make PRs with all the patches that have been sent.

nukeop:
How can I subscribe?
And how do we prevent spam and malicious entries?

shenlebantongying:
I think it's here, you can see my "test" email publicly, just click the "new post"

https://lists.sr.ht/~tyil/rms-support

Tyil:
> How can I subscribe?

There's a subscribe button, which should open your mailclient to send the appropriate mail to subscribe.

> And how do we prevent spam and malicious entries?

Spam is hard to prevent anywhere, even this Github repo can be spammed with frivolous issues. Malicious entries, however, could be stopped by requiring the patch to be signed, hence I ask  for signed patches.

For the record, this Github repo doesn't stop from making fraudulous entries either. I can make a new account, claim I am whoever I want to be, and submit a PR. I would argue we should focus on the positives, and not on potential issues.

nukeop:
Definitely, I'm just worried about potential abuse, because I have very little free time as it is. Thank you for the help. I will merge it.

PR #2605, opened 2021-03-26
Create ivakorin

ivakorin:


ivakorin:
Signed

alefvanoon:
The file name should be like *.yaml
so change to ivakorinl.yaml
image

ivakorin:
> 
> 
> The file name should be like *.yaml
> so change to ivakorinl.yaml
> image
Fixed
Thank you!

PR #2649, opened 2021-03-26
Update signed.html

cndkhuong:
Added Vietnamese


cndkhuong:
Added Vietnam translated menu

PR #2665, opened 2021-03-26
support RMS

nipinium:


trufanov-nok:
Please change extension to yaml instead of yml

PR #2677, opened 2021-03-26
Create QingGO.yaml

QingGo:

### DIFF
--- a/_data/signed/QingGO.yaml
+++ b/_data/signed/QingGO.yaml
@@ -1,0 +1,2 @@
+name: 曾颖清 (QingGO)
+link: https://github.com/QingGo
--- a/_data/signed/QingGo.yaml
+++ b/_data/signed/QingGo.yaml
@@ -1,2 +1,2 @@
-name: 曾颖清 (QingGO)
+name: Zeng Yingqing
 link: https://github.com/QingGo


_data/signed/QingGO.yaml
```diff
@@ -0,0 +1,2 @@
+name: 曾颖清 (QingGO)
```
tmwstw7: Hi there, I suppose your name in Chines fails CI\CD. I'm not trying to harass or smth. but currently everybody get this error
image

Lesik: In that case in my opinion the appropriate solution is to fix the CI script, not change the name.

Lesik: (Both yaml and Jekyll fully support Unicode.)

QingGo: I find that my PR did not pass the check. so i try to use PinYin to write my name in this PR, or maybe the check script should by modify.

tmwstw7: I think @Lesik is wright, there might be some problems with CI. Also I double-checked the verification script, it seems to work fine.

QingGo: glad to hear that

PR #2697, opened 2021-03-26
Updated Dutch translation

teunissenstefan:
Added Dutch translation to signed.html and signed_rtl.html, and translated README_NL.md


imachug:
Please revert signed.html and signed_rtl.html -- they're the same for every language. We'll fix the 'Signed' part a little bit later

imachug:
Thanks!

PR #2766, opened 2021-03-26
languages dropdown prototype

corporateanon:
- [ ] - check the accesibility
- [x] - fix the size


corporateanon:
Peek 2021-03-26 13-50

shenlebantongying:
Do we really have to introduce a  bunch of JavaScript...
I think the current one is quite cute. Anyone clicks the link then they can see a lot of flags for their language. They know that this is a true international thing and there are many of us across the global.

FSF also have a freeJS campaigns
https://www.fsf.org/campaigns/freejs/
https://www.gnu.org/philosophy/javascript-trap.en.html

image

corporateanon:
Got rid of JS

Peek 2021-03-26 23-20

corporateanon:
outdated

PR #2768, opened 2021-03-26
Add link to Tagalog translation

jobbautista9:
The Tagalog translation works, but it currently can't be found in the webpage. This PR fixes that.


trufanov-nok:
I guess link should be added to signed.html file too

jobbautista9:
You're right, I added the link to signed.html. The Vietnamese link should also be added to signed_rtl.html, but that needs to be done outside of this PR.

imachug:
Thanks!

PR #2778, opened 2021-03-26
Add signature

Stell0:
Free speech is freedom to say even bullshit.
End cancel culture.


6r1d:
Thank you for your support, you are right!

PR #2790, opened 2021-03-26
Create GitMensch.yaml

GitMensch:
no affiliation, because I only speak for myself, not for the projects I contribute to (or also lead, as this is a personal stand, not one I can make or anyone that contributes to my projects)


6r1d:
Thanks a lot!

PR #2793, opened 2021-03-26
AnnatarHe support rms

AnnatarHe:
support rms


trufanov-nok:
I guess the file should has .yaml extension, not .yml

PR #2959, opened 2021-03-26
Fixed bug in name by removing hyphen

lunar-mining:


shenlebantongying:
Your fork is too old. We now add new sign via .yaml file.

imachug:
Your fork is too old so I committed this change for you here: 31b1874e.

Issue #175, opened 2021-03-24
continuation: rms-open-letter denial of discussion


Jan200101:
continuation of https://github.com/rms-open-letter/rms-open-letter.github.io/issues/848 (issue has since been deleted) in a place where discussion is allowed
since the people behind that repository have shown they are unwilling to have one like an adult

Here is a list of locked issues
Issue | deleted | reason | notes | archive
------- | --------- | ---------- | ------- | -----------
https://github.com/rms-open-letter/rms-open-letter.github.io/issues/120 | no | too heated | responded after followup locking | https://archive.is/1lJKy
https://github.com/rms-open-letter/rms-open-letter.github.io/issues/401 | no | no reason | [2] | https://archive.is/990bb
https://github.com/rms-open-letter/rms-open-letter.github.io/issues/667 | yes | no reason | [1] linked
https://github.com/rms-open-letter/rms-open-letter.github.io/issues/672 | yes | no reason
https://github.com/rms-open-letter/rms-open-letter.github.io/issues/676 | yes | no reason | |  https://web.archive.org/web/20210324100146/https://github.com/rms-open-letter/rms-open-letter.github.io/issues/676
https://github.com/rms-open-letter/rms-open-letter.github.io/issues/696 | yes | no reason
https://github.com/rms-open-letter/rms-open-letter.github.io/issues/756 | yes | no reason
https://github.com/rms-open-letter/rms-open-letter.github.io/issues/794 | yes | no reason | [1] linked | https://archive.is/CM9C1
https://github.com/rms-open-letter/rms-open-letter.github.io/issues/811 | yes | too heated | [1] linked
https://github.com/rms-open-letter/rms-open-letter.github.io/issues/848 | yes | no reason | `I have no interest in discussing this further here.` - neilmcgovern
https://github.com/rms-open-letter/rms-open-letter.github.io/issues/909 | no | off topic | | https://archive.is/jnH0H
https://github.com/rms-open-letter/rms-open-letter.github.io/issues/924 | yes | spam
https://github.com/rms-open-letter/rms-open-letter.github.io/issues/939 | no | off topic | asking about a gitlab repository | https://archive.is/tOj7D
https://github.com/rms-open-letter/rms-open-letter.github.io/issues/940 | yes | no reason | asked where to support RMS | https://archive.is/A0yFZ
https://github.com/rms-open-letter/rms-open-letter.github.io/issues/949 | yes | too heated | | https://archive.is/i0lSW
[...]

PR #2962, opened 2021-03-26
handicraftsman's signature

handicraftsman:
I do not like what hapened, and i believe what RMS has done isn't completely okay, but who is flawless with today's western education and culture? As to me, many other events in western countries cause a worse cultural shock than what RMS has said.

What comes to RMS's specific case, what did you expect when new US federal and state-local laws regarding child sexuality are a thing?

This is definitely an issue, - not with RMS, but with US internal stuff.

Another point is that nevertheless RMS is a qualified programmer, who became who he is right now thanks to his skills too.

As such, I sign this letter to support RMS, and as a suggestion to discuss the issue in general, not applied specifically to RMS.

### DIFF
--- a/_data/signed/handicraftsman.yaml
+++ b/_data/signed/handicraftsman.yaml
@@ -1,0 +1,2 @@
+name: handicraftsman
+link: https://github.com/handicraftsman


handicraftsman:
Another issue is this - it means that the other side of this conflict isn't playing fair. If you don't know about this - the issue pointed to by the link got deleted by repository owners. They didn't want to discuss the question.

http://web.archive.org/web/20210326090023/https://github.com/rms-open-letter/rms-open-letter.github.io/issues/2250

shenlebantongying:
@handicraftsman 
There is list what they have censored if you may interest
https://github.com/rms-support-letter/rms-support-letter.github.io/issues/175

PR #3009, opened 2021-03-26
Better translation workflow(imo)

cmd410:
So, kinda new, more structural workflow for translations.

Write translation to `_translation/index_lang.md` where `lang` is a language code, add `emoji` attribute, and voila, it's shown on the page and available under `/index-lang.html` link, no other modifications required.

1. Unclutters root directory
2. Simplifies adding links to translations
3. Avoids merge conflicts when several translations add link to the same line

Example `_translations/index_de.md`:
```md
---
layout: signed
title: An open letter in support of Richard M. Stallman
description: An open letter in support of Richard Matthew Stallman being reinstated by the Free Software Foundation
image: /assets/social-media-preview.png
locale: de_DE
twitter:
  card: summary_large_image
emoji: 🇩🇪
---

2021-03-23

Richard M. Stallman, ...
```

Also, added translation contributing guidelines.

PR #3052, opened 2021-03-26
Fix flags isolation problem

shenlebantongying:
See last message of https://github.com/rms-support-letter/rms-support-letter.github.io/pull/3009

Its now 

image


shenlebantongying:
@imachug 
image

imachug:
Fixed, thx

PR #3162, opened 2021-03-26
Debloat footer

kitsunyan:
This patch removes repeating text from footer.

Screenshot_2021-03-27_00-15-41


imachug:
Makes sense, thanks.

PR #3165, opened 2021-03-26
Improve translations block style

kitsunyan:
This patch removes manual `<br>` insert after 11 flags, which fixes layout on mobile devices.

It also applies 50% width limit on medium and large screens (similarly to how it looks now in non-mobile browsers) and improves paddings.

P.S. With current 11-flags-delimeter they may be displayed like this on mobile devices (or even worse):

```
XXXXXXXXX
       XX
     XXXX
```

### DIFF
--- a/_includes/translations.html
+++ b/_includes/translations.html
@@ -8,8 +8,5 @@
       {%- assign c2 = langcode | slice: 1, 1 -%}
       {{- site.data.emoji[c1] -}}{{- site.data.emoji[c2] -}}
     </a>
-    {% if forloop.index == 11 %}
-      <br>
-    {% endif %}
   {% endfor %}
 </div>
--- a/_sass/custom.scss
+++ b/_sass/custom.scss
@@ -5,9 +5,22 @@
 
 .translations {
 	text-align: end;
+	margin: 0 0 1em 0;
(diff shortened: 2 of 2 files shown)


imachug:
Looks much better now, thanks!

shenlebantongying:
Sorry for my ad-hoc solution, but it was actually try to solve a problem

https://github.com/rms-support-letter/rms-support-letter.github.io/pull/3052

shenlebantongying:
Sry for my ad-hoc solution, but it actually tries to solve a problem that 1 or 2 flags maybe got isolated.
https://github.com/rms-support-letter/rms-support-letter.github.io/pull/3052

Maybe reduce its width to make it like this? I have no idea how css works.

image

Its more artistic than a straight line plus two isolated flags. Or maybe just reduce line space?

image

kitsunyan:
I don't see a big problem here. The amount of flags may be increased even more later, and I find it strange to adjust width every time the amount changes. Moreover, the amount of flags may be indivisible by 2/3/4/5 (let's assume we won't have more rows).

I understand some people may be obsessive about that (like I am about other things), but I don't see "good" solutions for this.

PR #3169, opened 2021-03-26
yeah

0x0000ff:
yeah


imachug:
Can you please replace 'Example name (Good company)' with your real name? Thanks!~

0x0000ff:
I'm sorry. I don't feel comfortable outright posting my real name on a public forum.

Vaporeon:
An alias/handle is still ok, as long as it does not say "Example"

0x0000ff:
OK sorted. Thanks.

Issue #2441, opened 2021-03-26
Extra page for the good RMS and the FSF have done?


ossimalvin:
Since the other repo has an extra page detailing why they believe RMS is evil incarnate, I think there should be an extra page on this repo detailing what the FSF has done for open source and maybe some links to donation pages and other ways to support the project.

shenlebantongying:
- https://jorgemorais.gitlab.io/justice-for-rms/
- https://sterling-archermedes.github.io
- https://www.wetheweb.org/post/cancel-we-the-web

6r1d:
[Proof RMS is Still Important](https://open.lbry.com/@Lunduke:e/RMSStillImportant:6?r=GAaJ9s3EZNYqp61oaAXUcUzgEMPmhhf1)

[Why RMS is a Great Mascot and spokesperson for Free Software
](https://open.lbry.com/@Lunduke:e/RMSGoodMascot:b?r=GAaJ9s3EZNYqp61oaAXUcUzgEMPmhhf1)

nukeop:
Please document this, it's very important.

dodobyte:
that's a good one
https://geoff.greer.fm/2019/09/30/in-defense-of-richard-stallman/

Maciejfiedler:
> Since the other repo has an extra page detailing why they believe RMS is evil incarnate, I think there should be an extra page on this repo detailing what the FSF has done for open source and maybe some links to donation pages and other ways to support the project.

Yeah this guy literally changed the whole goddamn world with GNU, GNU/Linux and FSF.

michaelkarlcoleman:
RMS has arguably done more to make the world a better place than any other computer programmer, living or dead.

jobbautista9:
We can put the extras to the repo's wiki, so that it can be easily edited.

shenlebantongying:
I think a Redditor has done a big job
https://www.reddit.com/r/freesoftware/comments/meot29/dissecting_hate_speech_the_rms_open_letter/

romanrm:
See https://stallmansupport.org/ and specifically https://stallmansupport.org/richard-stallman-honors-and-awards.html

shenlebantongying:
https://stallmansupport.org/ has been endorsed by RMS himself and that page link to us.

![image](https://user-images.githubusercontent.com/20123683/114263380-1f676e80-99b3-11eb-84b6-dddbb4ecf5dc.png)

[...]

PR #3219, opened 2021-03-26
Create ryanrms.yaml

ryanrms:
Adding signature in support of rms, I disagree with much of what he has to say but 100% support him as an ambassador of free software.


6r1d:
Hello! Many of the things he is criticized for aren't even his words. He did change his opinion on some things, as well. There are some [relevant links](https://github.com/rms-support-letter/rms-support-letter.github.io/issues/2441).

Let's look why `.yaml` is broken now...

Issue #2193, opened 2021-03-25
Can't see the flag in chrome on LUbuntu 20?


JiaoXianjun:
Hello,

I can see the flags in firefox on Ubuntu and with my iPhone chrome, but not in the chrome on Ubuntu. My chrome on Ubuntu is like this:
![screen](https://user-images.githubusercontent.com/5212105/112532547-3b5ef380-8da9-11eb-90a1-0abe010f4990.jpg)

imachug:
The site uses emojis. Perhaps something's wrong with fonts.

cmd410:
![](https://user-images.githubusercontent.com/48290258/112533902-808c2100-8dbb-11eb-860c-c1bd9b0bab76.png)

Wait, there are supposed to be flags..?

cmd410:
Windows 10, Mozilla. Same problem.

RealJTG:
Hehe, the reverse side of emojis. 
Mb we should include emoji webfont like https://cdnjs.com/libraries/emojione or https://github.com/googlefonts/noto-emoji?

@JiaoXianjun try `sudo apt install fonts-noto-color-emoji`

shenlebantongying:
yes..........

![image](https://user-images.githubusercontent.com/20123683/112534059-06d74180-8d81-11eb-91c5-c9dd27060ab7.png)

shenlebantongying:
I think we better add a font that contains emoji for those flags.

A few distro's default font config doesn't have it.............

Since this going to be access from people all over the world, we should add it directly to the repo.

shenlebantongying:
Or, we add some svg pictures of those flags?

JiaoXianjun:
> sudo apt install fonts-noto-color-emoji

sudo apt install fonts-noto-color-emoji -- it works for me: chrome on Ubuntu (mine is LUbuntu)

nukeop:
Do you guys know how to link the emoji font in Jekyll without using a cdn? Can anyone open a pull request?

RealJTG:
Although I've never used Jekyll before, there is https://github.com/jekyll/jemoji plugin, this probably will require replacing raw emojis with github placeholders like :us: -> `:us:`

RealJTG:
```diff
diff --git a/_config.yml b/_config.yml
index 2e15f6a4..fd85e2eb 100644
--- a/_config.yml
+++ b/_config.yml
[...]

PR #3240, opened 2021-03-27
Shrink NotoColorEmoji.ttf

kitsunyan:
This font is loaded every time the page is opened, and its size is 10MB. It's noticeable on slow connection.

Moreover, this font doesn't work on Firefox (probably related to [this issue](https://bugzilla.mozilla.org/show_bug.cgi?id=1454152)), but still it will try to download it.

So I decided to shrink the font, leaving only country flags:

```bash
pyftsubset 'NotoColorEmoji.ttf' \
  --output-file='NotoColorEmojiFlags.woff' \
  --layout-features='ccmp' \
  --flavor='woff' \
  --unicodes='U+1F1E6-1F1FF'
```

Also related to https://github.com/rms-support-letter/rms-support-letter.github.io/issues/2193.


shenlebantongying:
WOW, never know you can do that. I just leaned something `:)`

imachug:
Thanks a lot! I thought about shrinking the font too but somehow forgot about that.

PR #3243, opened 2021-03-27
Create JesseCharlie.yaml

JesseCharlie:


6r1d:
Uhh, these sites don't exist and there's a redirect to the White House on str.ong?! Can you at least use a real profile?

JesseCharlie:
Ok, how am I supposed to prove I am not a bot?

PR #3244, opened 2021-03-27
Create jjquin.yaml

jjquin:
I Disapprove of What You Say, But I Will Defend to the Death Your Right to Say It.
- Evelyn Beatrice Hall, 1906


6r1d:
Thank you for your support!

PR #3267, opened 2021-03-27
Create 1-7-1.yaml

1-7-1:
RMS is a pioneer of the free software movement and philosophy as well as a legendary programmer. He is a historical figure and his legacy speaks for itself. I don't agree with everything he's ever said, but when it comes to free software, he's always been on the right track and his words are still spot on to this very day. To try and get rid of Richard Stallman is a perverse act of historic revisionism and an thinly veiled attack on an icon of liberty around the world.

### DIFF
--- a/_data/signed/1-7-1.yaml
+++ b/_data/signed/1-7-1.yaml
@@ -1,0 +1,2 @@
+name: 171
+link: https://171.life/


6r1d:
Thank you for your support! By the way, while it is not required, adding a real name improves the credibility of the letter.

PR #3281, opened 2021-03-27
Add my signature (aaronfranke)

aaronfranke:
This PR adds my signature as `aaronfranke.yml`.

Additionally, I digitally signed the `index.md` file with my GPG key (can be verified with `gpg --verify`). It seems fitting that a letter in support of RMS is signed using a GNU tool. I placed the signature in a new `_/data/gpg` folder. Please let me know if there is somewhere else that you would prefer these files be placed.


imachug:
Hi, a GPG signature is a great idea, but I think it'd be easier for us all if you GPG-signed the commit itself. It's described here: https://git-scm.com/book/en/v2/Git-Tools-Signing-Your-Work. See 'GPG Introduction' and 'Signing Comments'.

aaronfranke:
@imachug The commit itself is also signed (it says "Verified"), but I figured I would include a signature of the text directly, so that the signature can be viewed without looking through the Git history.

imachug:
Ah, I didn't see that it's signed. Merging then!

PR #3339, opened 2021-03-27
Change Leah Rowe entry to say Leah Rowe (Libreboot Project Leader)

githubisnonfree:
Also changed URL from osboot.org to libreboot.org. I would rather link
Libreboot right now because that's a more well-known project of mine.


shenlebantongying:
WOW! That's unexpected :)

PR #3351, opened 2021-03-27
Richard M. Stallman

fintechrecruiters:

### DIFF
--- a/_data/signed/<fintechrecruiters>.yaml (replace <Stackedsats>
+++ b/_data/signed/<fintechrecruiters>.yaml (replace <Stackedsats>
@@ -1,0 +1,2 @@
+name:Stackedsats
+link:https://github.com/orgs/StackedSats


jobbautista9 (changes requested):
The filename is malformed. Also, please add a space in each field.

6r1d:
Hello and thank you for your support!

There's a number of things our parser doesn't understand, though. You don't need to name a file with `<>` for starters, `fintechrecruiters.yaml` or `Stackedsats.yaml` is ok. Is it "fintechrecruiters" or "Stackedsats"?
A space is missing after 'name:' in `<fintechrecruiters>.yaml`, so please add it.
A space is missing after 'link:' in `<fintechrecruiters>.yaml`, so please add it.
Stackedsats uses a strange protocol 'ttps' on line 2. Please use `https://` or `http://`.

6r1d:
Fixed in #3357

PR #3408, opened 2021-03-27
Add JP social media image to /assets, edit /_translations/index_jp.md to link to it

ChanoSan:


imachug:
I don't like that the text was moved very much to the left and it's no longer aligned with header. Do you know if you can fix that?

ChanoSan:
I shunted it over because Stallman's name was starting to cover his face. I can play with the font and size to try and get it back, let me fiddle with Inkscape a bit

ChanoSan:
I adjusted the font sizes so it stays in the same alignment as the EN image, also fixed the naming convention to social-media-preview_ja_JP

imachug:
Great, thanks a lot!

PR #3428, opened 2021-03-27
consistency of README filenames

jobbautista9:
per Discord convo with @imachug
done properly now


6r1d:
Will URLs for the files be auto-updated, too? The flag icon ones.

jobbautista9:
I don't think so. This change should only affect the READMEs, which don't appear in the webpage.

6r1d:
Alright, thanks :-)

PR #3435, opened 2021-03-27
Make file formatting POSIX compliant

aaronfranke:
This PR makes file formatting POSIX compliant, including by removing carriage returns (CRLF -> LF aka Windows format to Unix format) and ensure files end with newlines (required by POSIX definition of a text file). I also added a `.gitattributes` file which should prevent Git from letting any more carriage returns into the repo. Also, a BOM was removed in `README_PL.md`.


imachug:
There have been many new signatures since your PR, so I applied the changes manually. Thanks!

PR #3452, opened 2021-03-27
Create Palmay.yaml

Palmay:


6r1d:
Thank you for your support! 
Do you mind using a real name for the general credibility of the letter? If not, it's your choice. You can PR it later. :-)

PR #3464, opened 2021-03-27
Count on top

### LINKED ISSUE #3420: Put the total number of supporter at top?
Hello,

Each time when I try to check the number I have to scroll down to the bottom after long operation. Should we put the number at top of the page? Then people could see the number and fell encouraged easier?

Thanks.

cmd410:
Displays signatories count on top of the list like so:
изображение

Closes #3420


6r1d:
Thank you very much! :-)

JiaoXianjun:
When will we see the change about putting the total number at the top? At least, now there is nothing.

image

cmd410:
@JiaoXianjun it is right above the list of signatures

PR #3467, opened 2021-03-27
Add signature for verisimilitude

Tyil:
As requested on IRC, and stated on their personal website: http://verisimilitudes.net/rms-support


6r1d:
Thank you for your support!

PR #3489, opened 2021-03-27
Create pavel-krivanek

pavel-krivanek:


6r1d:
Hello, thanks for your support, it means a lot. And also, I really like the game you made, Pharo Chip Designer looks awesome!

PR #3500, opened 2021-03-27
Sorts signs in case-insensitive order.

wxy:
Sorts signs in case-insensitive order.


6r1d:
Thank you for your help :-)

PR #3501, opened 2021-03-27
Create ThinkPadThink

ThinkPadThink:


6r1d:
Isn't that a copy of #3504?

jobbautista9:
It is. Closing since #3504 is already merged.

PR #3517, opened 2021-03-27
Create franpoli.yaml

franpoli:
Support the FSF and RMS


6r1d:
Thank you for your support!

PR #3529, opened 2021-03-27
Signed the RMS support open letter

jkonecny12:
I may disagree with what he is saying but I definitely don't agree that everyone is attacking him these days. He dedicated most of his life for Open Software so we could everyone benefit from that. He made our world much better place! He don't deserve to be treated like this.


6r1d:
Thank you for your support, I appreciate it even more seeing what you do!

PR #3543, opened 2021-03-27
Create vitrvvivs.yaml

vitrvvivs:
Those attempting to push RMS out of the world he nearly-single-handedly created show their bigotry towards the neurodivergent. We should treat those who so obviously think differently with charity and kindness, not with anger. Everyone has to learn the rules of what is socially acceptable, and everyone learns at a different pace. For those on the spectrum, it can genuinely take 60 years to learn what others consider basic.


shenlebantongying:
thx

PR #3606, opened 2021-03-27
Create cactus06.yaml

cactus06:


6r1d:
That's an interesting list of repos to say the least.
Thank you for your support!

cactus06:
>That's an interesting list of repos to say the least.
Thank you for your support!

Are you a real person? Did it work properly?

6r1d:
I'm a real person. I think it'll appear soon.

PR #3607, opened 2021-03-27
Replace image in index_tl.md with Tagalog translation

jobbautista9:
Forgot to do that when I added the Tagalog translation of the social media preview.


imachug:
Thanks!

PR #3614, opened 2021-03-27
Create adokitkat.yaml

adokitkat:


6r1d:
Thank you for your support!

adokitkat:
> Thank you for your support!

I may have some problems with Richards personal opinions, however they have nothing to do with his professional life and I don't think he should be punished just for having different thoughts when not hurting anybody... That's not a right thing to do.

6r1d:
By the way, he writes a lot and changes his opinions in arguments. It's best to ask for a source if someone says "Richard told X".

PR #3649, opened 2021-03-27
Create ramshit.yaml

ramshiitt:


6r1d:
Do we accept emails as identifiers? Isn't it better to use GitHub, anyway?

jobbautista9:
Yes. But they should be prepended with mailto:

PR #3756, opened 2021-03-27
Create wiredancer.yaml

wiredancer:


shenlebantongying:
Your link doesn't seems to be valid, Is there a typo or something?

wiredancer:
The link works here. What is wrong with it?

shenlebantongying:
I cannot load it via firefox but chrome is ok.

PR #3762, opened 2021-03-27
Added social media preview in Spanish

cidoku:
Please tell me if anything else should be changed. If a more formal Spanish is required, I'll make the proper edits.


shenlebantongying:
I don't think it has to be very formal, since it mostly shared via social media.
 (I don't understand Spanish, so I will trust you)

`:)`

PR #3844, opened 2021-03-27
Create encw.yaml

encw:
I don't at all agree with everything Stallman has said, or done, but I think he deserves a second chance.


shenlebantongying:
Thanks!

PR #3850, opened 2021-03-27
Update french translation

franpoli:
### French translation update


shenlebantongying:
The img looks great. Thank you for updating!

Issue #3714, opened 2021-03-27
Translations updates needed


gorshunovr:
Hello everyone. We need support and PRs with translations of:
- `README_XX.md` files – need to be synced with English `README.md` version of the file (request for company in examples on lines 6 & 14)
- `assets/social_media_preview*.png` pictures. Source of picture: [link](https://github.com/rms-support-letter/rms-support-letter.github.io/files/6205876/rms-design-source-v2.zip).
  Corresponding `_translations/index_XX.md` file needs to refer to newly uploaded picture in its header.

Calling on @rldmt @cidoku @mahdyar @ahangarha @doctorx32 @Tyil @Bluesboy @bnainar @jobbautista9 @McSinyx.
Will highlight more names later.

@imachug please, label 'needs help'.

Thank you!

McSinyx:
It's night time here, I'll be updating the Vietnamese translation 
tomorrow (I'm on phone atm).

imachug:
I've added the Russian version

Bluesboy:
@imachug so my assistance don't needed anymore?

imachug:
Please check if you like it and if you do--no.

![social_media_preview_ru_RU](https://user-images.githubusercontent.com/16370781/112731077-3daa8480-8f46-11eb-94a6-cd601edd9145.png)

rldmt:
@imachug how did you edit the picture ?

Tyil:
The Dutch version of the README still looks fine from what I can see, same goes for the `nl_NL` version of the picture. Is there anything I'm missing, or could the Dutch versions be up to date already?

Bluesboy:
Added links to non-Github signing to Russian translation.

xgqt:
I want to make a Polish "social-media-preview_pl_PL.png"
Is there a cleared version of it or am I blind again

teunissenstefan:
> The Dutch version of the README still looks fine from what I can see, same goes for the `nl_NL` version of the picture. Is there anything I'm missing, or could the Dutch versions be up to date already?

I updated them like 8 hours ago.

bnainar:
How do I edit the image? @gorshunovr @imachug

jobbautista9:
@bnainar See #3324.

jobbautista9:
@Tyil The Dutch translation seems to be missing the video link.

teunissenstefan:
@jobbautista9 Updated

bnainar:
[...]

PR #3865, opened 2021-03-27
Update Korean translation

greenhorse-effect:


greenhorse-effect:
This update corresponds with #3714

shenlebantongying:
Since you originally translated it. I will merge it directly `:)` 

The pic looks good,

PR #3879, opened 2021-03-28
Create AlfredWingate.yaml

parona-source:
I am signing in support of RMS. He may be rude and pedantic at times, but that doesn't justify forching him to resign nor resignation of the whole board of the FSF.


shenlebantongying:
Thanks!

Issue #3867, opened 2021-03-27
Indicating FSF associated members here?


shenlebantongying:
One can donate to FSF directly to become an associated member

Donating directly to FSF may help the situation.

For existing members, maybe we should add affiliation like  `name (FSF member)`

https://www.fsf.org/associate/

FrostKnight:
That is actually a good idea.

cmpunches:
I would consider donating but without receiving associated member status.  

I stay completely unburdened by authority or external influence by policy.  This allows me to embrace the concepts that work for me while still being able to reject the strings that often come with them.  I would enjoy those types of associations but most of the problem-people in the f/oss space like to attack peoples' relationships as parts of their strategy, so, I have to treat any links as liabilities.

aleksejrs:
> I would consider donating but without receiving associated member status.

https://my.fsf.org/donate

franpoli:
An [FSF board FAQ](https://www.fsf.org/news/fsf-board-frequently-asked-questions-faq) was released yesterday with information about financial status.

>The vast majority of FSF’s financial support comes from individuals -- many, but not all, of whom choose to become associate members. At this moment, the FSF has more associate members than at any time in its history.

This indicates we are not giving in to intimidation and that the public rejects the smear campaign. We should maintain this trend in order to emphasize that we do not accept discrimination and defamation.

PR #3906, opened 2021-03-28
Update italian and french transaltion

franpoli:
#### What does this implement/fix?

Update the Italian Translation with the associated image.

Minor correction of the French translation and update of the associated image with the same phrasing


franpoli:
As suggested on issue #3867, I am informing that I am an FSF associated member.

6r1d:
I've read your updates using the translator. Thank you for your help. It's nice to see we'll have more credibility to support RMS :-)

PR #3993, opened 2021-03-28
Create 1ykos.yaml

1ykos:
Let's work together even if our viewpoints don't align perfectly so long as we have just one shared goal, and in this case free software.


jobbautista9:
Thanks! :D

PR #3997, opened 2021-03-28
Create bonktree.yaml

bonktree:

### DIFF
--- a/_data/signed/bonktree.yaml
+++ b/_data/signed/bonktree.yaml
@@ -1,0 +1,2 @@
+name: Arseny Maslennikov
+link: mailto:arseny@altlinux.org


shenlebantongying:
Do you also want to add affiliation of alt linux?

bonktree:
> Do you also want to add affiliation of alt linux?

I have signed my commit with a GPG key to avoid signature forgery, so I'd prefer the yaml file to stay as it is.
As for the mention in the auto-maintained signatory list — I don't really care if my ALT affiliation is mentioned there, the maintainers know best how to do it. Whether it's one way or the other. I'm fine with it.

shenlebantongying:
ok

PR #4024, opened 2021-03-28
Signed gladly. Stand with Stallman.

lordfeck:
Stand with Stallman and software freedom.


6r1d:
Thank you for your support!

lordfeck:
You're most welcome @6r1d. The more voices the better.

PR #4051, opened 2021-03-28
Update README_UA.md

banderlog:
"Alternative way" and codeberg vote path added. And a 'i stand with EMS' picture


bnainar:
Which language?

banderlog:
Ukrainian. Same as it was before for this file.

PR #4095, opened 2021-03-28
My signature to support Richard Matthew Stallman

andrcmdr:
Without Richard Matthew Stallman (aka RMS) there wouldn't be GNU project, without GNU there wouldn't be GNU/Linux and Linux kernel would be useless, without RMS and GNU there wouldn't be Free/Libre and Open Source Software communities and organizations as we know them - we must to remember the commitment of Richard Matthew Stallman for his groundbreaking work on GNU, we must be grateful and appreciate RMS for the Freedom in Software which we have today!


shenlebantongying:
Thanks!

PR #4110, opened 2021-03-28
Create aldo-herese.yaml

Aldo-Herese:
This resented MOB is extremely destructive for our community, and don't forget, they are always a minority, a very tiny minority.  Shame on us if we let them lead us by the nose


shenlebantongying:
Thanks!

PR #4136, opened 2021-03-28
Create skeletonium.yaml

skeletonium:
Add my signature to support Richard Stallman against reprehensible and unjustified attacks from the cancel culture mob.


shenlebantongying:
Thanks!

PR #2065, opened 2021-03-25
Update legolivesmatter.yaml

LegoLivesMatter:

PR #4211, opened 2021-03-28
Add notice about not representing the whole Void

Logarithmus:
See https://github.com/rms-support-letter/rms-support-letter.github.io/pull/2065#issuecomment-808973364

### DIFF
--- a/_data/signed/Logarithmus.yaml
+++ b/_data/signed/Logarithmus.yaml
@@ -1,2 +1,2 @@
-name: Artur Sinila (Void Linux contributor)
+name: Artur Sinila (Void Linux contributor, but I don't represent the whole organisation)
 link: https://logarithmus.dev


shenlebantongying:
OK.

shenlebantongying:
You must be very important for VoidLinux i guess. Thanks for supporting us <3

Logarithmus:
> You must be very important for VoidLinux i guess. Thanks for supporting us <3

Not as important as @q66 (his contributions are an order of magnitude more then mine), but thank you.

Duncaen:
You don't represent the organization at all.

sgn:
Please remove the whole affilication. The part `doesn't represent the whole organization` partly implies that you're represent part of the organization, which is false.

shenlebantongying:
Please be kind to each other since you guys are working on the same project.

PR #4250, opened 2021-03-29
Create lierdakil.yaml

lierdakil:
I don't have any particular love for RMS -- let's face it, he's more than a bit eccentric -- but this "cancelling" campaign which quotes the man out of context and goes out of its way to vilify him and hide or downplay every good thing RMS has done makes me pretty sick.


shenlebantongying:
Thanks

PR #4257, opened 2021-03-29
Screw up scrapper

powerboat9:
Screws up the scrapper [here](https://gist.github.com/kyhwana/73161a50d1b5019b18e3965b55a1a192) using a [spider trap](https://www.fleiner.com/bots/). Should screw up some other curl based scrappers.

### DIFF
--- a/_includes/signatures.html
+++ b/_includes/signatures.html
@@ -1,5 +1,6 @@
 <p><b>Signed by {{ site.data.signed.size }} individuals:</b></p>
 <ol>
+  <!--<li><a href="https://github.com/[1-999][1-999][1-999][1-999]">Spider Trap</a></li>-->
 {%- capture signs -%}
   {%- for sign_hash in site.data.signed -%}
     {%- assign signature = sign_hash[1] -%}


shenlebantongying:
Approved!

shenlebantongying:
Some Hilarious bullshits for some tracking to this pull request.
https://archive.is/1bHIq

nukeop:
Can you inject `rm -rf ~/`? Or a fork bomb?

powerboat9:
> Can you inject `rm -rf ~/`? Or a fork bomb?

I don't think so. The way xargs is used, the only thing you can really do is screw with the url parameter curl gets. You could try using a name starting with "../../" to call a different api function, though you'd have to find an api function that takes the same parameters in a POST request.

nukeop:
They seem to have adapted to this hack in a fork...

powerboat9:
#4661

powerboat9:
#4667

nukeop:
Won't that break some links? Like `mailto:`?

shenlebantongying:
yes, reverted

https://github.com/rms-support-letter/rms-support-letter.github.io/commit/6ad7901bd456ef6ae3feeff939d59d54102291da

nukeop:
Thanks for fixing.

nukeop:
Looking at that scraper, we could add a fake attribute to each `<a>` tag, e.g. `<a herp="https://github.com/derp\>" href=[real link]` and it would fuck it up again because it looks for everything between the beginning of github url and `\>`.

PR #4299, opened 2021-03-29
Direct link to directory

nin-jin:


jobbautista9:
Thanks, do you mind if you sync the Russian translation with the current version of the README?

nin-jin:
EN version currently have direct link.

jobbautista9:
Yeah, but it also has a link to a video containing visual instructions, which the RU version currently doesn't have.

PR #4339, opened 2021-03-29
Create 0x0000_dot_ru.yaml

0x0000-dot-ru:


6r1d:
Hello, thanks for your support, your url doesn't work, and GitHub profile can add a bit more credibility. Update it if you don't mind, please. Thank you!

PR #4368, opened 2021-03-29
Updated Italian Translation

dag7dev:


jobbautista9:
Great, LGTM via machine translation. Can you also consider syncing the translation with the current index.md? Thanks!

dag7dev:
> Great, LGTM via machine translation. Can you also consider syncing the translation with the current index.md? Thanks!
Machine translation? Last commit has been done taking consideration of some users which suggested me more suitable words.
Anyway, I will update index.md too in case of newer untranslated things, I've already translated it some days ago.

PR #4397, opened 2021-03-29
Remove my sign

DanInSpace104:
I don't want anymore to be part of any side in this conflict. In any conflict to be honest. I think it is against my life-vision, and choosing side was wrong decision. Please, remove my sign.


DanInSpace104:
@6r1d Thank you, and good luck for you too

PR #4443, opened 2021-03-29
Create dieterweber.yaml

uellue:
See https://www.wetheweb.org/post/cancel-we-the-web, in particular the comment from Nadine Strossen. I couldn't have said it better and fully agree.

If one agrees with what RMS says or not, if one likes RMS or not, that's a different topic. What I'd like in a community:

* Be kind and tolerant
* Assume good intentions
* Be objective and follow due process
* Hard on the problem, soft on the people

This can be hard. Everybody has already violated these guidelines, including me. We all need each other's help achieving these goals. https://en.wikipedia.org/wiki/Jesus_and_the_woman_taken_in_adultery

The comments and actions from some people who call for the removal of RMS are worse than what RMS himself has said and done, as far as it is documented. Everybody should take a good look in the mirror, tone down, and find a way to work together or at least not wage war against each other. Being kind and tolerant towards people we like and we agree with is easy. Being kind and tolerant towards people we dislike and disagree with is the hard part.

### DIFF
--- a/_data/signed/dieterweber.yaml
+++ b/_data/signed/dieterweber.yaml
@@ -1,0 +1,2 @@
+name: Dieter Weber
+link: http://materials-scientist.com/


jobbautista9:
Thank you for the support! :D

PR #4549, opened 2021-03-29
Add NomisIV

NomisIV:
I don't have to support everything Richard Stallman has ever done, to support his work in the FSF.


shenlebantongying:
Thanks!

PR #4557, opened 2021-03-29
Create eaty.yaml

ertanatay:
FSF and RMS must not yield to the mob! Free Software is freedom in the modern world.


shenlebantongying:
Thanks!

PR #4615, opened 2021-03-29
Add Hungarian translation

friendlyanon:


shenlebantongying:
Thanks!

No idea about the language, so i will trust you `:)`

friendlyanon:
You may use https://www.deepl.com/translator to double-check translations.  
It is quite more reliable than other services when it comes to languages that are either difficult to translate (like Japanese) or spoken by less people (like Hungarian).

PR #4645, opened 2021-03-29
Create Skilvingr.yaml

Skilvingr:
There is not much to say: FSF is Richard Stallman's creature and curse it for what he said, whatever he said, I beg judgmentals's pardon, is simply a crime against a human being. Cancel Culture is damaging the entire world, dragging down people who express their thoughts unclearly, and, paradoxically, hiding out the real causes of problems. Please, stop this, at least in the name of the free software.


shenlebantongying:
Thanks!

PR #4667, opened 2021-03-29
Actually break scrapper

powerboat9:
Makes previous round of scrapper breaking actually work. Turns out, you need '/[', not just '[', in the url.

### DIFF
--- a/_includes/signatures.html
+++ b/_includes/signatures.html
@@ -23,7 +23,7 @@
   {%- assign name = sign_items[2] -%}
   {%- assign link = sign_items[4] -%}
   {% comment %}Add whitespace{% endcomment %}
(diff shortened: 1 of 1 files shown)


shenlebantongying:
As long as it doesn't break the links, we are good.

shenlebantongying:
https://gist.github.com/kyhwana/73161a50d1b5019b18e3965b55a1a192/revisions

That guy updated the script, does its still works though?

shenlebantongying:
If grep only works line by line, I hink i can alter our html table to 1 line.

shenlebantongying:
Hahahah, yes. That script now become shit again `:)`

nukeop:
Excellent 👍🏿

PR #4703, opened 2021-03-29
Added Galician translation

martinge17:


_translations/index_gal.md
```diff
@@ -0,0 +1,71 @@
+---
+layout: signed
+title: Unha carta aberta en apoio a Richard M. Stallman
+description: Unha carta aberta en apoio a reincorporación de Richard M. Stallman á Free Software Foundation
+image: /assets/social-media-preview_gal_GAL.png
+locale: gal_GAL
```
shenlebantongying, line 6: Is gal_GAL a vliad locale name?

shenlebantongying:
I dont actually know your language. Does this flag the right one?
image

_translations/index_gal.md
martinge17: I think it is gal_ES but I can't confirm that

martinge17:
galicia-clipart-4

No, this is the flag.

shenlebantongying:
I am so sorry, but it seems Galician's flag doesn't be implemented yet for emoji..............
https://github.com/Crissov/unicode-proposals/issues/330
https://www.google.com/get/noto/help/emoji/flags/

shenlebantongying:
For some reason,  gal_GAL and ES-GA are mapped to this flag
is there a political issue here?

image

martinge17:
No, I don't think there is a political issue, thank you.

_translations/index_gal.md
polenta-e-osei: Is `<lang_code>_<country_code>`, since the Galicia is not recognized as a country but as a simple region of Spain (with some degree of autonomy, but is not relevant), also the correct `lang_code` is `gl`, so it should be `gl_ES`.
For example, if I remember correctly, the locale for Catalan as `ca_ES` is widely used.

Source: [localeplanet.com](https://www.localeplanet.com/icu/)

PR #4739, opened 2021-03-30
Create hispachan_t.yaml

CuloArdido:
I saw that this letter is being discussed on https://www.hispachan.org/t/res/81493.html and someone proposed the idea of signing on behalf of the board.


shenlebantongying:
WOW, thanks a lot. Since signing our letter is suiciding for big organizations, you guys are the No.1 organization that sign this `:)`

If individuals want to sign, they can still feel free to sign.

PR #4745, opened 2021-03-30
Create hiljusti.yaml

hiljusti:
I don't agree with all RMS's views, and I don't believe the way he expresses them is always helpful or appropriate.

That said, he's a reasonable human, he learns from mistakes, and makes changes. I believe in the power of forgiveness. I don't believe in the power of banishment.


shenlebantongying:
Thanks!

PR #4911, opened 2021-03-30
Update 404.html

zemuro:


shenlebantongying:
Why you make this change? Does it relate to mobile device?

PR #4913, opened 2021-03-30
Adds a couple newlines to break scraper scripts

JackRedstonia:
Self-explanatory code and PR title.
I might add a couple more mechanisms to prevent scraping in the future, who knows. Have fun with this one for now.

### DIFF
--- a/_includes/signatures.html
+++ b/_includes/signatures.html
@@ -22,7 +22,9 @@
   {%- if sign_items[2] and sign_items[4] -%}
   {%- assign name = sign_items[2] -%}
   {%- assign link = sign_items[4] -%}
-  <li><a href="{{ link }}">{{ name }}</a></li>
+  <li>
+  	<a href="{{ link }}">{{ name }}</a>
+  </li>
   {%- endif -%}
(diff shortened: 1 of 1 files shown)


McSinyx:
But why should we prevent scraping?  This is a political statement and I doubt anyone here is ashamed of having it known to the public.

JackRedstonia:
Think of it as an insurance policy, it simply prevents scrapers from using the letter as a vector for targeted harassment.

shenlebantongying:
Which script are you mentioning?

I put all <li> in one line to break this one https://gist.github.com/kyhwana/73161a50d1b5019b18e3965b55a1a192

I can also track other ones.

JackRedstonia:
@shenlebantongying I think it's the script you mentioned. Does this PR break it again?

shenlebantongying:
No, just two different method to break it. Idc, it seems the original author already lost his interest.

There is some fun bullshit :)
https://archive.is/1bHIq

JackRedstonia:
> It seems the original author already lost his interest.

It's still necessary to weed out the script kiddies.

PR #4948, opened 2021-03-30
Create dnscrypt.pl

pysiak:


luukasahola:
Please change the filename into `dnscrypt.yaml` or `dnscrypt.pl.yaml` for the system to recognize the signature.

PR #5012, opened 2021-03-30
Sending power ot RMS

foss-respecter:


shenlebantongying:
Sorry, we don't accept this kind of sign, since we are getting malicious signs.

Issue #3991, opened 2021-03-28
Add Armenian Translation


appetrosyan:
Hi. There's a large FOSS community in a small country, and even though an Armenian localisation is wholly unnecessary (in that most people who can code, can read English, and sometimes well), it may be helpful to boost morale.

jobbautista9:
Sure! PRs welcome.

bnainar:
Done yet?

appetrosyan:
Ah,  sorry, I haven't even started yet.

jobbautista9:
Take your time :)

PR #5065, opened 2021-03-31
Armenian translation

appetrosyan:
Should address the first part of #3991.


jobbautista9:
Google doesn't translate Armenian well, it seems. I will need some help from another native Armenian speaker.

appetrosyan:
@jobbautista9 Google is an ML-based translation engine, it has a habit of adding things that often follow in newspapers. This is a hand translation. I'm looking for people in my immediate vicinity that could help with double-checking the translation.

jobbautista9:
Ok, I've also asked for help in IRC, Matrix, and Discord, if that helps. Someone in IRC said they know someone who speaks Armenian somewhat.

appetrosyan:
A friend of mine also agreed to help (and maybe sign), but only in 7 hours or so.

jobbautista9:
Since the translation of index.md is not yet done, I've marked this PR as a draft for now. Feel free to mark as ready when you're done translating.

appetrosyan:
AFAIK, I'm supposed to create a file with the right localisation in `_localisations`.

jobbautista9:
Ah yes, I missed those commits, thanks for marking the PR as ready.

shenlebantongying:
If we cannot find a reviewer, I think we better just merge it.

Once it got wider audiences, someone else would correct mistakes.

shenlebantongying:
It looks good.
image

PR #5236, opened 2021-03-31
Added sharing links.

Wafelack:


shenlebantongying:
Thanks for adding that.

However, it seems Reddit and Twitter are unnecessary now. They are none-free platform (in contrast to mastodon/pleroma), and this letter already shared everywhere so many times (almost everyone within the OSS community already knew it).

Wafelack:
Ok I understand, sorry for the disturbance.

PR #5247, opened 2021-03-31
Add Jan Hamal Dvořák

mordae:
Signing after reading this: https://www.wetheweb.org/post/cancel-we-the-web


shenlebantongying:
Thanks!

PR #5253, opened 2021-03-31
Added hreflang

taiju:
Added hreflang so that Google can crawl multilingual support effectively.

I tried to respect the language and not limit the region, because to reach a wide range of users, but I am a little lost. For example, for English, I specified `en` instead of `en-US`. (However, the flag emoji is the United States.)

This is a proposal, so I'll leave it to you to decide.

ref:
https://developers.google.com/search/docs/advanced/crawling/localized-versions
https://en.wikipedia.org/wiki/List_of_ISO_639-1_codes


shenlebantongying:
@nukeop

imachug:
OK for now but the list should be generated automatically

imachug:
Nah, I'm an idealist :)

shenlebantongying:
Too real :)
https://lists.gnu.org/archive/html/emacs-devel/2018-03/msg00089.html
image

taiju:
It is better to generate them from the file names, but some of them are incompatible with ISO 639-1.
For example, the Albanian language uses `al` for file names, but the ISO 639-1 notation is `sq`.

taiju:
Correction, it is better to generate it from the Front Matter locale. However, it is difficult to automate the process if the file names are not consistent.

imachug:
> It is better to generate them from the file names, but some of them are incompatible with ISO 639-1.

There's `locale` field in the pages headers.

taiju:
Yes, as I corrected earlier, I also thought it would be better to base it on the `locale` field in the header (the `locale` field in Front Matter). 

Front Matter https://jekyllrb.com/docs/front-matter/

imachug:
Oh, sorry, I had no idea it's called Front Matter.

taiju:
Sorry for the lack of explanation here as well :bow:

I'll try to automate the process tomorrow, along with fixing the file name, as it will be troublesome to manage.

PR #5331, opened 2021-03-31
Create apatil

apatil:
I am signing because, as far as I know, Stallman is accused of: 1. having a bed in his office, 2. being despondent at one point and wrongly attempting to pressure a woman into dating him, 3. having unpopular opinions, which he has arrived at in good faith and with good will, and which he has been willing to reconsider as he has learned new things. I am signing despite the risk to my personal reputation and career because I want the world to be safe for awkward free thinkers who occasionally become despondent and do the wrong thing. Also, no one alive today knows what the final endpoint of human morality will look like, when the arc of history finishes bending toward justice. Our history up to now should teach us humility. It's overwhelmingly likely that even the most progressive among us have thoughts and behaviors that our descendants will recognize as backward or barbaric, and that there exist oppressed groups to whose plight we are blind. People like Stallman, who are humane, principled and brave enough to point out reasonable objections to overwhelmingly popular norms and taboos, help us all get better more quickly.


shenlebantongying:
No worry, there are 5k+ retarded (positive meaning) with you signed this in the risk of personal reputation `:)`

apatil:
As far as I know, the term "retarded" is still considered pejorative by most people with cognitive disabilities (https://en.wikipedia.org/wiki/Retard_(pejorative)). I therefore urge you not to use it at all.

shenlebantongying:
ok

apatil:
Thank you for the encouragement!

PR #5371, opened 2021-03-31
Create jschwartzenberg.yaml

jschwartzenberg:
I went through the letters. The unfortunate letter against RMS writes "RMS recommends" and clicking the link behind it there is written "I think". I see a pattern of twisted words to try make a statement, but this is not an appropriate way to misrepresent someone even when that person has made make big mistakes otherwise. Addressing issues becomes harder when the circumstances are dishonest. The unfortunate letter creates an atmosphere in which people who have valid criticism or who have been mistreated, face extra difficulty to stand out and have their concerns addressed. I encourage independent concerns to be addressed as such and to avoid any association with the unfortunate letter which I'm fearing is already mistakenly being done by some supporters of this letter.

### DIFF
--- a/_data/signed/jschwartzenberg.yaml
+++ b/_data/signed/jschwartzenberg.yaml
@@ -1,1 +1,2 @@
 name: Julius Schwartzenberg (I cannot judge if RMS should stay on the FSF board or not, the twisting of words in the other letter is disturbing and helping neither cause.)
+link: https://github.com/jschwartzenberg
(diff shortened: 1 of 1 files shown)


shenlebantongying:
I agree with you.

However, can you add a link to your github profile or personal website?

shenlebantongying:
Thanks!

jschwartzenberg:
Ah it was not clear this was not associated that is added now.

Thank you for your remark. I really hope that others who have spoken out against RMS will not be associated with the unfortunate letter when no such link is put by them and their criticism might be valid. I believe RMS has a strong character and can handle a lot but if there are indeed vulnerable people with valid criticism, they should be listened to independently. It is an easy tendency to to associate it all together while other criticism that is out there does appear to have validity to it. I hope we can make it easier for everybody through this letter.

shenlebantongying:
Sure

PR #5404, opened 2021-04-01
Improving PT-BR translation

Naranbataar:


shenlebantongying:
Thanks!

After seeing `_data / assinado / <nomeutilizador> .yaml`, i suspect the last translator just use google translator................

PR #5484, opened 2021-04-01
Update README_IT.md

ostendali:


6r1d:
Hello. Are you sure that linking yourself in a Readme is the best idea? How about someone neutral, like John Smith, but for italian language?

ostendali:
hi there, it is not really clear how to signup for the petition and I think most of people are having the same difficulty and therefore are not signing....you will need to come with something better where they just put their names and email addresses maybe...

shenlebantongying:
Done

PR #5512, opened 2021-04-01
Modified my profile

taiju:
I became a Associate Member as from today!

ref #3867


6r1d:
Thank you for your support! By any chance, did you write to FSF that your reason for becoming an associate member is to support RMS? :-)

taiju:
@6r1d Yes! I answered "RMS open letter" to the question "What inspired you to join today?" when I registered.

PR #5513, opened 2021-04-01
use native words alongside english ones

drage0:
the document should use as many native words as possible.
some technical jargon such as "pull request" and "fork" can't be translated directly, i have tried to translate them as best as i can for the native speaker. (in addition, ascii and forking wikipedia article have been linked)
the changes also include a few structural differences that i believe read better.


marksbizness:
Dynin Mark

PR #5531, opened 2021-04-01
Automated hreflang generation

taiju:
rel #5253

I tried to automate the generation of hreflang, but there were a lot of exceptions to handle, which increased the complexity.
I'll leave it to you to decide whether to adopt it or not.

@imachug


imachug:
You don't have to generate the URL directly with `https://rms-support-letter.github.io/{{filename_prefix}}{{downcase_hreflang}}{{filename_suffix}}`. You can use `{{ translation.url }}` instead, like we do in _includes/translations.html.

taiju:
@imachug Cool!!

_includes/hreflangs.html
```diff
@@ -1,33 +1,6 @@
...
-<link rel="alternate" hreflang="tr" href="https://rms-support-letter.github.io/index-tr.html" />
-<link rel="alternate" hreflang="uk" href="https://rms-support-letter.github.io/index-ua.html" />
-<link rel="alternate" hreflang="vi" href="https://rms-support-letter.github.io/index-vi.html" />
-<link rel="alternate" hreflang="zh-Hans" href="https://rms-support-letter.github.io/index-zh-cn.html" />
-<link rel="alternate" hreflang="zh-Hant" href="https://rms-support-letter.github.io/index-zh-tw.html" />
-<link rel="alternate" hreflang="en" href="https://rms-support-letter.github.io/" />
-<link rel="alternate" hreflang="x-default" href="https://rms-support-letter.github.io/" />
+{% for translation in site.translations %}
+  {%- assign hreflang = translation.url | replace: '/index-', '' | replace: '.html', '' -%}
+  <link rel="alternate" hreflang="{{hreflang}}" href="https://rms-support-letter.github.io{{translation.url}}" />
```
imachug, line 3: IIRC `href` doesn't have to be absolute, so we don't need the https://rms-support-letter.github.io prefix

taiju: I am referring to [this URL](https://developers.google.com/search/docs/advanced/crawling/localized-versions#all-method-guidelines), which contains the following information.

> Alternate URLs must be fully-qualified, including the transport method (http/https), so:
> https://example.com/foo, not //example.com/foo or /foo

imachug: Oh. It's probably better to use `{{ translation.url | absolute_url }}` then though.

taiju: Thanks! I fixed it with 7026475.

imachug:
Thanks!

PR #5535, opened 2021-04-01
Fixed to keep line breaks

taiju:
Fixed to keep the line breaks in the README, referring to the GitHub Flavored Markdown Spec.

If this is what you intended, please withdraw it.

ref: [6.13 Soft line breaks](https://github.github.com/gfm/#soft-line-breaks)

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,33 +1,33 @@
 # An open letter in support of RMS.
 
-This README is available in:
-[🇦🇱](README_AL.md)
-[🇦🇪](README_AR.md)
-[🇩🇪](README_DE.md)
-[🇪🇸](README_ES.md)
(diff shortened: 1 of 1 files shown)


taiju:
A preview of the modified version can be seen at the following URL
https://github.com/rms-support-letter/rms-support-letter.github.io/pull/5535/files?short_path=b335630#diff-b335630551682c19a781 afebcf4d07bf978fb1f8ac04c6bf87428ed5106870f5

imachug:
The flags occupy the whole screen this way, so I think what he have now is intended. Thanks anyways!

taiju:
Thanks for the confirmation. I have no objection!

PR #5586, opened 2021-04-01
Delete AlexBethel.yaml

AlexBethel:
My decision to sign this document last week was an impulsive one I made under peer pressure and without the prerequisite knowledge or time to make an informed decision. Having had ample time to do research and consider the situation, the fact is I'm just not comfortable offering my unilateral support to this letter.

I would therefore like to respectfully request that my signature be removed from the list. My sincere apologies for wasting everyone's time.


shenlebantongying:
No worries, have a good day!

PR #5659, opened 2021-04-02
Create SyrusAkbary.yaml

syrusakbary:
While I don't endorse Richard Stallman point of view on many (many) aspects, I believe it's healthy to defend the right to have other perspectives even though they might not match our own. I don't see punishment and retaliation as optimal ways to make this world a happier place.

I defend the right to make mistakes as a healthy way of evolving our own understanding.

As such, I'm signing this note from a personal manner (not my employer or company) :)


6r1d:
Thank you for your support! It's great to see you with us.

PR #5687, opened 2021-04-02
capitalised first letter

charlie39:


6r1d:

charlie39:
> Thank you for your support! Please consider sharing this letter on the social media.

sure, i have already shared it in a couple of linux facebook groups, and will share where i could. Thank you

PR #5740, opened 2021-04-03
Add signature for Yaroslav de la Peña Smirnov

Tyil:
https://lists.sr.ht/~tyil/rms-support/patches/21750


6r1d:
Oh, thank you for transfering this signature. It helps!

PR #5775, opened 2021-04-03
Create notpushkin.yaml

notpushkin:


6r1d:
Hello. Thanks for your support, it helps a lot. If you can, share the letter on social media and add projects you participate in.

PR #5868, opened 2021-04-03
Indicate FSF membership

Hhina:
As suggested in <https://github.com/rms-support-letter/rms-support-letter.github.io/issues/3867> and <https://libreboot.org/news/rms.html>.


shenlebantongying:
Thanks!

PR #5966, opened 2021-04-04
Create punarinta.yaml

punarinta:


6r1d:
Thank you for your support! It's very nice to see organizations with us, as well, because the facts are spreading.
It's hard to completely agree with someone, but the way Richard is was treated by open-letter is just wrong.
What matters these days, at least in my opinion, is to think critically, separate facts from someone's ideas and make sense of what's happening. And many do that in issues and discussions.
The whole anti-RMS campaign might separate the community for a long time, and this should not be forgotten.
Please consider sharing this letter on the social media, as well.

PR #5988, opened 2021-04-04
Create Bolshevikov

Bolshevikov:


6r1d:

6r1d:
In the end, we had a signature duplication. Fixed in https://github.com/rms-support-letter/rms-support-letter.github.io/commit/37f22ba9c2d9e87382fba9a59e15f970c50cbfff

There's your [signature](https://github.com/rms-support-letter/rms-support-letter.github.io/blob/master/_data/signed/Igor.Bolshevikov.yaml).

PR #6141, opened 2021-04-06
stop cyber-bullying

thread13:
An appalling example of cancel culture getting toxic.

Let us have a civilized discussion -- with facts, arguments, and hearing all parties involved -- instead of a plain witch-hunt, shall we?


6r1d:
Thank you for your support! We tried to establish a discussion [there](https://github.com/rms-support-letter/rms-support-letter.github.io/issues/175), many, many times.
For now, please consider sharing this letter on the social media.

PR #6146, opened 2021-04-06
Create earthengine.yaml

earthengine:
I have not check what RMS said, but it is RIDICULOUS to remove a well known person that have great contribution to FSF and the open source software development.


jobbautista9:
Thanks!

PR #6156, opened 2021-04-06
Bullies under disguise are still obvious

oelmekki:
Normies are very sad they're not allowed anymore to be racists, sexists,
and many other aspects of hate, so they are targeting the next "weird"
guy. "There's no label on that one yet, right? Great! Bring the
stones!". Nope, you're still dicks.


jobbautista9:
Thanks!

PR #6246, opened 2021-04-06
Create memclutter.md

memclutter:


6r1d:
Hello. Thank you for your support! By the way, the format was `.yaml`, not `.md`, but I fixed that.
Please consider sharing this letter on the social media.

PR #6284, opened 2021-04-06
Create y_s_shevchenko.yaml

lazy-starwing:
Denounce the witch-hunt against RMS and the FSF


6r1d:

PR #6292, opened 2021-04-06
Create christian-leger.yaml

christian-leger:
Note: I originally attempted to email my signature as a patch to signrms@prog.cf on April 2nd (subject: Please add Christian Leger to list; attachment filename: Create-cleger.yaml.patch). If it's still in the queue, please disregard it and use this pull request instead.


6r1d:

PR #6299, opened 2021-04-06
Create logic-debugger.yaml

logic-debugger:


6r1d:
I have merged your signature in https://github.com/rms-support-letter/rms-support-letter.github.io/pull/6298 already, don't worry.

PR #6305, opened 2021-04-06
Create kilianpg.yaml

kilianpg:


6r1d:
By the way, isn't FSFE well, [less than enthusiastic](https://fsfe.org/news/2021/news-20210324-01.en.html) about RMS being in FSF?

kilianpg:
Thanks.

Aspie96:
Hi, @kilianpg!

If I may, I suggest you to support the FSF rather than the FSFE (and not to support the FSFE).

The FSF has Stallman on the board. The FSFE wants him out of the board.

PR #6420, opened 2021-04-07
Sign the open letter

esanchma:
I still remember how Jake was removed from TOR. This whole thing proves that ESR was right.


shenlebantongying:
Thanks!

PR #6439, opened 2021-04-07
Create c3d.yaml

c3d:
Signing in my personal capacity.
My employer is not aligned with this support.


6r1d:
Thank you! Well, I just hope you'll stand your ground if somebody tells you to change your mind without arguments, there's some evidence supporting RMS here.

c3d:
Thanks. I investigated the evidence before signing.

PR #6452, opened 2021-04-07
Add me to support RMS

moonburnt:
its completely unacceptable to bully someone, regardless of difference in your personal opinions. Those who started campaign against grandpa of opensource should be ashamed


6r1d:

PR #6559, opened 2021-04-09
Create Mascantosh.yaml

Mascantosh:

PR #6562, opened 2021-04-09
Remove signature

alecks:
I've decided to remove my signature after realising that most of the signatories and parts of the letter represent a political stance, rather than RMS's right to remain in his position regardless of his beliefs.


shenlebantongying:
https://github.com/rms-support-letter/rms-support-letter.github.io/pull/6559

PR #6563, opened 2021-04-09
Create jemiapo.yaml

jemiapo:
name: Serge Jemia
link: https://play.google.com/store/apps/details?id=io.pokerplatform


shenlebantongying:
Your link doesn't work and your account is too new.

jemiapo:
Yea!
Above projet is available for early access now only.
Just see screenshot 
image
It will be open later.
So how old my account should be to accept my voice?

shenlebantongying:
Thanks for support!

PR #6569, opened 2021-04-09
add subscription

eBenya:


shenlebantongying:
Why you change so much files?

6r1d:
The surname changes for Yura. From "Tkachenko" to "Emelyanenko". Or closely sounding one. It is an edit, not an addition. Same with Eugeny... Nah, none of this makes any sense.

PR #6582, opened 2021-04-09
Create lvazquez.yaml

lvazquez:
I totally agree with the concepts in "An open letter in support of Richard M. Stallman". 
Even without agreeing with a good part of its controversial expressions, I consider freedom of expression and thought a supreme principle to be defended. The differences that arise from it must be debated in that field; without the use of extortion, without ignoring the context and without the use of the straw man fallacy as tools to silence, bastardize or punish different views, no matter how controversial they may be.


shenlebantongying:
Thanks!

PR #6636, opened 2021-04-11
Create ciastex.yaml

Ciastex:
How about we focus on what the person has achieved in their lifetime that we benefit from nowadays, instead of looking at their personal opinions. If FSF reinstated RMS they likely had a good reason to do so. Cancel """culture""" is pathetic and unfair. 

"Dangerous force" my ass.

### DIFF
--- a/_data/signed/ciastex.yaml
+++ b/_data/signed/ciastex.yaml
@@ -1,0 +1,2 @@
+name: Tomasz Cichoń
+link: https://github.com/Ciastex


6r1d:
Thanks for your support. Yeah, I agree with you, but many of opinions were fabricated, taken out of context or changed. We had a ton of discussions about those in the issues. Besides that, yes, RMS had acheived quite a bit in regards to licensing and his work is currently very important.

PR #6637, opened 2021-04-11
Create csardelacal.yaml

Csardelacal:
I'd like to extend my support to the FSF and RMS. Not because I agree with the views of RMS, but because I think that:

 1. We should not deliver mob justice to anyone. If somebody does horrible things they should be investigated by the authorities in charge.
 2. Their opinions, religion or politics do not improve or worsen their merit as a contributor to free software.

### DIFF
--- a/_data/signed/csardelacal.yaml
+++ b/_data/signed/csardelacal.yaml
@@ -1,0 +1,2 @@
+name: Cesar de la Cal (M3W)
+link: https://twitter.com/Csardelacal


6r1d:
Hi there! Thanks for your support.

> Not because I agree with the views of RMS

Just please don't assume all of his views are equal to what an open letter described. We've tried to look into it extensively in the repo's issues and found a lot of different info, often different from what an open letter claims.

Csardelacal:
Sorry, I meant it more in a "I don't agree with everything he has to say /thinks, but I value his opinions and I don't think anybody should be cancelled over views that may be controversial". I didn't assume the open-letter to be absolutely true, nor do I expect the support-letter to deal with absolute truths.

My judgement just lead me to consider the cause in the support-letter to be "more true" and less vile. Thanks @6r1d for making the letter and standing up to the mob!

PR #6640, opened 2021-04-11
Create PowerLay

PowerLay:


6r1d:
For stats: merged in https://github.com/rms-support-letter/rms-support-letter.github.io/pull/6639

PR #6651, opened 2021-04-11
matrix gateway

kompowiec:
finally IRC is the central point for bridges. Good work bros.

### DIFF
--- a/README.md
+++ b/README.md
@@ -60,8 +60,8 @@ If you still require help via visual instructions, use [this](https://invidious.
 
 ## Chatrooms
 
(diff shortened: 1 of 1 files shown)


imachug:
I don't think it's a good idea. The bridges are as follows ATM:

- Discord #general-1 on its own
- Matrix #general-1 on its own
- Jabber #support-rms on its own
- Discord #general-2 <-> Matrix #general-2 <-> Telegram @free_rms
- Discord #press <-> Matrix #press
- Discord #random <-> Matrix #random
- Discord #repo-discussion <-> Matrix #repo-discussion
- Matrix #irc <-> IRC #free-rms <-> Jabber bridge

Most stuff is available on Matrix, except local #general-1 channels, while IRC is only connected to a Jabber room and a local room. Matrix is still the largest hub.

kompowiec:
Well, I think it is worth combing than to divide. IRC is a good starting point for bridging the other protocols because:

- is simple protocol and easy to implemented
- works everywhere
- it's a IFTC standard
- backward compatibility - my xmpp server don't use biboumi just something else but still works. Meanwhile bridge with matrix to IRC break occasionally due to protocol changes

jobbautista9:
I think this will have to go through an issue first, before we merge this PR.

PR #6687, opened 2021-04-12
Create lottev1991.yaml

lottev1991:
I know RMS already stays on board, but after reading his very sincere post on the FSF website, I've hereby granted my definite support.


6r1d:

PR #6732, opened 2021-04-13
Create vbwagner.yaml

vbwagner:


6r1d:
Hi! I remember visiting your LiveJournal! :-)
Thanks for supporting RMS at hard times. Please share the news about our letter if you can.

PR #6798, opened 2021-04-16
Update Ratniex.yaml

Ratniex:
Not sure if this is worth mentioning at all. Maybe this cannot be considered a lot. I do not know. It is barely more than the price of one gnu. Close this without merging if you feel this is not appropriate.


shenlebantongying:
Thanks man!

6r1d:
That's a nice number :-)

PR #6898, opened 2021-04-25
nb_NO translation

comradekingu:
It pains me to not remove Telegram and Discord.

Should be "GitHub" "Jabber" and "ASCII" in the original.

Added a link to the GPL-3.0-only license text.
Changed the text to ~"objective journalists".


6r1d:
> It pains me to not remove Telegram and Discord.

The good thing here is that people will be discussing better alternatives.

Issue #6078, opened 2021-04-05
What can we improve from now on?


6r1d:
Things are calming down, but we have two communities from now on. One denounces RMS, other doesn't.
There's still little communication and there's a bit of conflict when somebody doesn't want to listen to you.

The fact many don't know the situation can potentially hurt RMS at a later date.
FSF is almost broken, and it's probably broken because it relied on sponsorship money.
The fact our letter only had appeared only in some news is strange, as well.

What can we do from now on to mend and improve things? Should we act at all? I propose an open discussion on that matter.

Qeenon:
I have suggestion: let close all our opensource projects (sorry in advance)

6r1d:
> (sorry in advance)

You don't have to be sorry for helping us think. Nobody should feel this way. There are too much reaction to the very discussion, so we get distracted and point fingers. And that hurts the community.

> I have suggestion: let close all our opensource projects

You have a point, actually. But that way helps the large companies, because it'll lead at worst to development of better disassembler and more closed source future. Or only one community that tends to ignore evidence.

Qeenon:
I understand. That's my initiative and there is no alternatives for now. But that's something alive right now.

6r1d:
> I understand. That's my initiative

Yes, you decide.

> and there is no alternatives for now. But that's something alive right now.

Well, we have lots of good people with us. I think there are many things we can do.

shenlebantongying:
The mission of this repo is completed. Bumping the number up is no longer impactful. The purpose is **not** to "fight over number". It's meaningless.

> There's still little communication

No, there are shit ton of communication already. Besides, the message already spread to the whole community.

Further things don't have to initiate from this repo. Just wait for the next move of FSF. 

Instead, we should stop accepting new signs as the open letter, so that we can do something else.

6r1d:
> The mission of this repo is completed. Bumping the number up is no longer impactful. The purpose is not to "fight over number" or "fighting between two parties". It's meaningless.

Huh, you have an interesting point. And you already told me something along the lines before.

> Further things don't have to initiate from this repo. Just wait for the next move of FSF.

Makes sense.

> Instead, we should stop accepting new signs as the open letter, so that we can do something else.

Honestly, it feels like a good idea to me, too.

nukeop:
> Further things don't have to initiate from this repo. Just wait for the next move of FSF.
> 
[...]

PR #6914, opened 2021-04-27
Sourcehut patches

Tyil:
I personally don't think it's still needed to continue adding signatures, but I've been getting requests from a number of people across multiple channels. As such, I will continue to add signatures added through the mailing list option, granted that they are valid patches that I can apply cleanly on the repository.


franpoli:
The question was discussed on issue [6078](https://github.com/rms-support-letter/rms-support-letter.github.io/issues/6078#issuecomment-816706369). I presume that this decision could be revaluated.

PR #6989, opened 2021-04-29
Add Armenian to the list of translations

dmayilyan:
Add a flag and a link to the Armenian readme file.


jobbautista9:
Huh, I wonder how we missed that. Thanks!

PR #7030, opened 2021-05-06
Create rasteric.yaml

rasteric:


rasteric:
He has always reminded me of Socrates but maybe we can do better now than 2500 years ago.

6r1d:

PR #7068, opened 2021-05-23
Create F.V.S.yaml

comsenwar:


6r1d:

comsenwar:
Thanks a lot for the ability to support Richard Stallman in such way. The e-mail is correct, and I'd be glad to answer any question if you have.

PR #7079, opened 2021-05-26
Completely migrate from freenode to libera

jobbautista9:
New freenode management has hijacked our channel in freenode, even though we have been clear that we will stay neutral from this mess and accommodate users on both networks. But alas, freenode has now shown that they can't be trusted. I propose we complete this unfortunate migration to libera.chat.


imachug:
LGTM. Honestly I didn't expect them to do so, they should have contacted me before renaming the channel at least, it wasn't a priority.

imachug (approved):

PR #7144, opened 2021-06-24
Create 000exploit.yaml

000exploit:


6r1d:
Thank you for supporting Richard Stallman! Please consider sharing this letter on the social media.

PR #7159, opened 2021-07-03
Add signature for JFryy

JFryy:
Signature for rms support


6r1d:
Hello and thank you for supporting Richard Stallman. :-)

PR #7162, opened 2021-07-03
📝 Adds My Signature

SandroLinux:
Adds my signature


6r1d:
Thank you for supporting Richard Stallman!

PR #7166, opened 2021-07-05
Create shinesprite

shinespritewastaken:


shinespritewastaken:
I should probably comment this:

I am in NO WAY attempting to claim the actual owner of the name "shinesprite" on GitHub, I'm trying to signify [with the slash] that some sites I use (where the name "shinesprite" is not taken), I use that name on, such as Discord. (And I know Stallman doesn't support Discord :smile:)

6r1d:
Thank you for supporting Richard Stallman! :-)

PR #7170, opened 2021-07-08
Create Andy-Python-Programmer.yaml

Andy-Python-Programmer:
Let's work together even if our viewpoints don't align perfectly so long as we have just one common goal, and in our case free software.


6r1d:
Thank you for supporting Richard Stallman! Indeed, let's work together for the common goal. :-)

PR #7173, opened 2021-07-09
Create tomonacci.yaml

tomonacci:
Thank you for putting together this support letter!


6r1d:
Thank you for supporting Richard Stallman. :-)

PR #7205, opened 2021-11-16
add my signature + fix typo in readme

CalrethilOfMirkwood:

### DIFF
--- a/README.md
+++ b/README.md
@@ -60,7 +60,7 @@ If you want to support the letter **without using Github**, go here: https://cod
 
 If you still require help via visual instructions, use [this](https://invidious.snopyta.org/watch?v=1lz5S5oS8CU) video.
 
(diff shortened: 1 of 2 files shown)


shenlebantongying:
There is no typo.

CalrethilOfMirkwood:
Now that I read it again the situation is a bit ambiguous and both sentences technically are grammatically correct.

Adjective phrase: Would you like to remove your signature from (**the letter** _in support of Richard Stallman_)?
vs.
Adverb phrase: Would you like to **remove** your signature from (the letter) _against Richard Stallman_?

However, I would recommend taking my suggestion because the first one intuitively makes more sense and the second one can come off as aggressive.

shenlebantongying:
I am sorry, but I am not a native English speaker and the another reviewer isn't available.

